PR #66, opened 2015-01-13
Updated marc languages mapping from lcc source

billdueber:
The language code map that ships with Traject is either incomplete or out-of-date; it doesn't include a bunch of newer stuff (Croatian/Serbian is what triggered it for me). Updated with new map pulled from the LCC XML dump.

### DIFF
--- a/lib/translation_maps/marc_languages.yaml
+++ b/lib/translation_maps/marc_languages.yaml
@@ -1,102 +1,105 @@
-# Map Language Codes (in 008[35-37], 041) to User Friendly Term
-
-# ???: null
-aar: Afar
+---
 abk: Abkhaz
 ace: Achinese
 ach: Acoli
 ada: Adangme
 ady: Adygei
-afa: Afroasiatic (Other)
+aar: Afar
 afh: Afrihili (Artificial language)
 afr: Afrikaans
-ajm: Aljamia
+afa: Afroasiatic (Other)
+ain: Ainu
 aka: Akan
 akk: Akkadian
 alb: Albanian
 ale: Aleut
 alg: Algonquian (Other)
+ajm: Aljamía
+alt: Altai
+tut: Altaic (Other)
 amh: Amharic
-ang: English, Old (ca. 450-1100)
+anp: Angika
 apa: Apache languages
 ara: Arabic
+arg: Aragonese
 arc: Aramaic
-arg: Aragonese Spanish
-arm: Armenian
-arn: Mapuche
 arp: Arapaho
-art: Artificial (Other)
 arw: Arawak
+arm: Armenian
+rup: Aromanian
+art: Artificial (Other)
 asm: Assamese
-ast: Bable
 ath: Athapascan (Other)
 aus: Australian languages
+map: Austronesian (Other)
 ava: Avaric
 ave: Avestan
 awa: Awadhi
 aym: Aymara
 aze: Azerbaijani
-bad: Banda
-bai: Bamileke languages
-bak: Bashkir
+ast: Bable
+ban: Balinese
+bat: Baltic (Other)
 bal: Baluchi
 bam: Bambara
-ban: Balinese
-baq: Basque
+bai: Bamileke languages
+bad: Banda languages
+bnt: Bantu (Other)
 bas: Basa
-bat: Baltic (Other)
+bak: Bashkir
+baq: Basque
+btk: Batak
 bej: Beja
 bel: Belarusian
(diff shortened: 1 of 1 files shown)


billdueber:
Merging. Just put this here as documentation.

jrochkind:
thanks!  I forget if we already have an automated script to pull, we could potentially create one, and even make a rake task for it, or even have something that warns you if the included one is "too old" and you should check for an update.

billdueber:
The problem with an automated script is that currently the tmap is "locked"
inside the traject gem itself. So, we'd have to make sure people are
putting a new version of the file somewhere it's going to be found before
traject itself; I need to revisit that logic to see how you implemented it.

On Tue, Jan 13, 2015 at 1:22 PM, Jonathan Rochkind <notifications@github.com
> wrote:

> thanks! I forget if we already have an automated script to pull, we could
> potentially create one, and even make a rake task for it, or even have
> something that warns you if the included one is "too old" and you should
> check for an update.
>
> —
> <https://github.com/traject-project/traject/pull/66#issuecomment-69792463>
> .
>



-- 
Bill Dueber
Library Systems Programmer

jrochkind:
Oh yeah, I was thinking mainly about the automated script for use by we
traject devs.

But there is a way to make it so a local copy of the tmap in a certain
location will take precedence over the one in the traject gem -- in fact, I
think this may work already, there is a load path that traject uses to look
up tmap definitions, it will use the first one it finds, and the local app
file system paths should be first in the list before the gem file system
paths. I think it may do that already -- I constructed it with that sort of
thing in mind. But if it doesn't, it can be made to.

billdueber:
It *does* do that already -- I'm doing it right now -- so I think we're
good to go. I'll look for a good place to document it, too. I don't know
how often those files (language, geographic, etc.) change -- we may want to
take out the shipped versions, and throw an error saying, "Hey, if you want
to use that, you need to create a `lib/translation_maps` directory and run
`traject generate marc_language` or whatnot. But as I type that, it seems
horribly difficult to explain.

For the moment, maybe we'll just try to keep up and generate new point
releases when those files change.

On Tue, Jan 13, 2015 at 4:34 PM, Jonathan Rochkind <notifications@github.com
> wrote:

> Oh yeah, I was thinking mainly about the automated script for use by we
> traject devs.
>
> But there is a way to make it so a local copy of the tmap in a certain
> location will take precedence over the one in the traject gem -- in fact,
> I
> think this may work already, there is a load path that traject uses to
> look
> up tmap definitions, it will use the first one it finds, and the local app
> file system paths should be first in the list before the gem file system
> paths. I think it may do that already -- I constructed it with that sort
> of
> thing in mind. But if it doesn't, it can be made to.
>
> On Tue, Jan 13, 2015 at 1:47 PM, Bill Dueber <notifications@github.com>
> wrote:
>
> > The problem with an automated script is that currently the tmap is
> > "locked"
> > inside the traject gem itself. So, we'd have to make sure people are
> > putting a new version of the file somewhere it's going to be found
> before
> > traject itself; I need to revisit that logic to see how you implemented
> > it.
> >
> > On Tue, Jan 13, 2015 at 1:22 PM, Jonathan Rochkind <
> > notifications@github.com
> > > wrote:
> >
> > > thanks! I forget if we already have an automated script to pull, we
> > could
> > > potentially create one, and even make a rake task for it, or even have
> > > something that warns you if the included one is "too old" and you
> should
> > > check for an update.
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <
> > https://github.com/traject-project/traject/pull/66#issuecomment-69792463>
>
> > > .
> > >
> >
> >
> >
> > --
> > Bill Dueber
> > Library Systems Programmer
> > University of Michigan Library
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/traject-project/traject/pull/66#issuecomment-69796429>
> > .
> >
>
> —
> <https://github.com/traject-project/traject/pull/66#issuecomment-69824600>
> .
>



-- 
Bill Dueber
Library Systems Programmer

jrochkind:
Cool. I would definitely not want to take out the shipped versions -- making everyone personally and locally keep updating their mapping files is no benefit over having traject do it. Most people are no more likely to be more up to date themselves than traject is (and people can choose to do that if they want anyway).  Traject should do these things out of the box -- and for most people most of the time, being a bit out of date is no disaster. When someone notices, traject can be updated (once for everyone, instead of telling all users to do it for themselves separately!)

PR #88, opened 2015-02-13
Switch to using FastXMLWriter

billdueber:
Require fastxmlwriter in the specfile, use it to generate XML, and change the test to actually test round-tripness.

### DIFF
--- a/traject.gemspec
+++ b/traject.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "yell"                        # logging
   spec.add_dependency "dot-properties", ">= 0.1.1"  # reading java style .properties
   spec.add_dependency "httpclient", "~> 2.5"
+  spec.add_dependency 'marc-fastxmlwriter', '>=0.1' # fast marc->xml
   
   # If we're building the package under JRuby, add in the 
   # jruby-only gems and specify the platform.
--- a/traject.gemspec
+++ b/traject.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "yell"                        # logging
   spec.add_dependency "dot-properties", ">= 0.1.1"  # reading java style .properties
   spec.add_dependency "httpclient", "~> 2.5"
-  spec.add_dependency 'marc-fastxmlwriter', '>=0.1' # fast marc->xml
+  spec.add_dependency 'marc-fastxmlwriter', '~>1.0' # fast marc->xml
   
   # If we're building the package under JRuby, add in the 
   # jruby-only gems and specify the platform.
--- a/lib/traject/macros/marc21.rb
+++ b/lib/traject/macros/marc21.rb
@@ -3,6 +3,7 @@ require 'traject/translation_map'
 require 'traject/util'
 require 'base64'
 require 'json'
+require 'marc/fastxmlwriter'
 
 module Traject::Macros
   # Some of these may be generic for any MARC, but we haven't done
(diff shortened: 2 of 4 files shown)


billdueber:
If we decide to go this way, I'll throw out an rc2.

jrochkind:
It makes sense to me to go this way, what do you think? 

The only thing I'd do is make sure that all the tests (if any!) from ruby-marc testing XML serialization are also in marc-fastxmlwriter (or equivalent tests that test the same thing).

traject.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "yell"                        # logging
   spec.add_dependency "dot-properties", ">= 0.1.1"  # reading java style .properties
   spec.add_dependency "httpclient", "~> 2.5"
+  spec.add_dependency 'marc-fastxmlwriter', '>=0.1' # fast marc->xml
```
jrochkind: If we're releasing a traject 2.0 depending on it, do you think it's time to call marc-fastxmlwriter 1.0?  Then we could semver our dependency to `~> 1.0` (ie, 1.x). 

Otherwise, we should probably dependency on  `">= 0.1", "< 2"` for future-safeness, rather than any version at all greater than 0.1.

jrochkind: http://semver.org/

> How do I know when to release 1.0.0?

> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you're worrying a lot about backwards compatibility, you should probably already be 1.0.0.

billdueber:
+1 to releasing fastxmlwriter as 1.0. Will do that nowish.

As for tests, the ruby-marc tests are all against either specific elements, or look for a specific string and/or rexml object. What I'm doing now is taking all the (non-on-purpose-broken) marc files from the ruby-marc project and round-tripping them with FastXMLWriter and MARC::XMLReader and testing equivalence of the starting and ending record. I'm not willing to say it's provably perfect, but I've been using it for months and had no trouble.

billdueber:
Travis passes, I've been using it for a while, and @barmintor had good luck with his tests. Any objection to a 2.0.0.rc.2 release?

barmintor:
+1

jrochkind:
I mean, the ruby-marc tests _could_ of course be duplicated in fastxmlwriter, by using the same MARC sample records, right?

But if you're relatively confident that fastxmlwriter tests covers everything the ruby-marc XML serialization tests cover (which may not be very much), that's good enough for me, I trust your judgement. Just thought the ruby-marc tests covering this functionality would be a useful thing to look at to make sure test coverage, at least, is enough to be as confident in fastxmlwriter as in the ruby-marc func it's replacing.

jrochkind:
Out of curiosity, how did barmintor's performance go compared to SolrMarc when using fastxmlwriter, even under MRI?

(And I'm impressed you were able to identify the original bottleneck so quickly, Bill!)

billdueber:
@barmintor reported going from about 140r/s to about 520r/s. So about a
factor of 4.

On Fri, Feb 13, 2015 at 1:16 PM, Jonathan Rochkind <notifications@github.com
> wrote:

> Out of curiosity, how did barmintor's performance go compared to SolrMarc
> when using fastxmlwriter, even under MRI?
>
> (And I'm impressed you were able to identify the original bottleneck so
> quickly, Bill!)
>
> —
> <https://github.com/traject-project/traject/pull/88#issuecomment-74299578>
> .
>



-- 
Bill Dueber
Library Systems Programmer

jrochkind:
Even running traject under MRI? **SWEET**

billdueber:
@barmintor Released rc2. You can continue with what you've got, or go back to using 

```ruby

to_field 'marc_display',  serialize_marc(:format => :xml)

```

jrochkind:
Oh wait, the 140r/s is with slow-xml-serialization, not what he was getting with SolrMarc, right?

That's still awesome, but I'm curious about this real world example comparing SolrMarc performance to traject-even-under-MRI.

jrochkind:
Also, @Billdueber, it does seem like a fastxmlwriter option (or default, or only option) should be included directly in ruby-marc, eh?

billdueber:
He claimed that his original 120r/s was "about 3x slower" than solrmarc. So, even under MRI he's going faster. Although I'm not positive he was using the exact same indexing.

@barmintor, if you have an actual apples-to-apples comparison you can run and share, I would *love* to see those number.

PR #93, opened 2015-02-20
Default to production date, when available.

MrDys:
For records with a 'p' date_type, default to the second date (date of production). This is particularly useful for video/film resources where users will be trying to narrow results by release date, not the date the media carrier was produced.


billdueber:
Thanks @MrDys! Merging; will go into next release.

PR #95, opened 2015-02-23
Gc min

barmintor:
These changes were minimizing object creation hotspots indicated by StackProf

### DIFF
--- a/lib/traject/indexer.rb
+++ b/lib/traject/indexer.rb
@@ -161,7 +161,7 @@ class Traject::Indexer
   # to be called for each record, and generate values for a particular
   # output field.
   def to_field(field_name, aLambda = nil, &block)
-    @index_steps << ToFieldStep.new(field_name, aLambda, block, Traject::Util.extract_caller_location(caller.first) )
+    @index_steps << ToFieldStep.new(field_name.freeze, aLambda, block, Traject::Util.extract_caller_location(caller.first) )
   end
 
   # Part of DSL, register logic to be called for each record
@@ -568,15 +568,19 @@ class Traject::Indexer
 
     def execute(context)
       accumulator = []
-      [@lambda, @block].each do |aProc|
-        next unless aProc
-
-        if aProc.arity == 2
-          aProc.call(context.source_record, accumulator)
+      if @lambda
+        if @lambda.arity == 2
+          @lambda.call(context.source_record, accumulator)
         else
-          aProc.call(context.source_record, accumulator, context)
+          @lambda.call(context.source_record, accumulator, context)
         end
-
+      end
+      if @block
+        if @block.arity == 2
+          @block.call(context.source_record, accumulator)
+        else
+          @block.call(context.source_record, accumulator, context)
+        end
       end
       return accumulator
     end
--- a/lib/traject/marc_extractor.rb
+++ b/lib/traject/marc_extractor.rb
@@ -105,6 +105,11 @@ module Traject
   # lazily create and then re-use a MarcExtractor object with
   # particular initialization arguments.
   class MarcExtractor
+    # constant values
+    EMPTY = []
+    FIELD_880 = '880'.freeze
+    SUBFIELD_6 = '6'.freeze
+
     attr_accessor :options, :spec_hash
 
     # First arg is a specification for extraction of data from a MARC record.
@@ -146,7 +151,7 @@ module Traject
       if options[:alternate_script] != false
         @interesting_tags_hash['880'] = true
       end
-
+      @interesting_tags = @interesting_tags_hash.keys
       self.freeze
     end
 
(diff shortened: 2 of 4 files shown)


lib/traject/indexer.rb
```diff
@@ -568,15 +568,19 @@ def inspect
...
-          aProc.call(context.source_record, accumulator, context)
+          @lambda.call(context.source_record, accumulator, context)
+        end
+      end
+      if @block
+        if @block.arity == 2
+          @block.call(context.source_record, accumulator)
+        else
+          @block.call(context.source_record, accumulator, context)
         end
```
jrochkind: Huh, this un-DRYing has measured perf improvements, eh?

lib/traject/marc_extractor.rb
```diff
@@ -105,6 +105,11 @@ module Traject
   # lazily create and then re-use a MarcExtractor object with
   # particular initialization arguments.
   class MarcExtractor
+    # constant values
+    EMPTY = []
```
jrochkind: Better freeze it too. `EMPTY = [].freeze`.  Cause if anyone does accidentally modify it, it'll cause a world of hurt when everyone is sharing the same `EMPTY` array (not threadsafe either).

jrochkind: and some comments explaining this is done for performance and has measurable perf advantages wouldn't hurt either, so some future coder doesn't get rid of it without measuring.

lib/traject/marc_extractor.rb
```diff
@@ -264,17 +267,17 @@ def extract(marc_record)
     #
     # Third (optional) arg to block is self, the MarcExtractor object, useful for custom
     # implementations.
-    def each_matching_line(marc_record)
-      marc_record.fields(@interesting_tags_hash.keys).each do |field|
+    def each_matching_line(marc_record,tags=nil)
+      tags ||= @interesting_tags
+      marc_record.each_by_tag(tags) do |field|
```
jrochkind: each_matching_line is meant to be public API that you can call on a MarcExtractor to do custom things. I don't understand what this second arg `tags` is, it hasn't been documented, and I think it ruins it as API -- it's probably something the MarcExtractor already knows about, so doesn't make that much sense as API? It still works the same if called without this optional arg? That might be okay then. But it needs to be documented in comments what it is, and why it's been made an arg, and why you'd ever choose to pass in a second arg instead of leaving it the default.  There are a few other places in this PR

lib/traject/indexer.rb
jrochkind: Sorry to be adding more comments as I think of them -- if the bottleneck was array creation (which I think barmintor suggested in IRC), can we obtain the same speed benefits by putting the `[@lambda, @block]` array into a class constant like you've done in other places, instead of un-DRY unwrapping of the loop?

It would be nice to keep it DRY if possible -- features we're thinking of, like automatic profiling, might require changes at this point in the code, and keeping it DRY would avoid having to repeat the same changes twice too. (That particular feature might not end up being implemented at this code location, but thinking of it re-triggered my aversion to un-DRY'ing here, heheh). 

The change could be made at the analagous location [here](https://github.com/barmintor/traject/blob/gc-min/lib/traject/indexer.rb#L515) too. 

It would also be even more affirming if the change were benchmarked to provide significant improvements in MRI 2.2, not just 2.1, to further help overcome my aversion to making the code somewhat more confusing, demonstrating it really does provide advantage.

jrochkind: Ah wait, it makes no sense to put `[@lambda, @block]` into a global constant, since those are iVars of course. 

Hmm. Any other patterns we could use to avoid the array creation while keeping it DRY? A submethod instead? Something like:

    call_proc(@lambda)
    call_proc(@block)

Method call isn't prohibitively expensive, is it?  Sorry to be picky here, but I think us being picky and careful so far is part of what's led to traject being what it is. 

Do you have any scripts or tests you use for benchmarking/profiling you can share, so others like me or Bill can try alternatives and see if they perform just as well with cleaner code? (Might apply to future too, when someone looks to alter or refactor some of this code, and wants to make sure they arent' performance regressing your changes).

barmintor: I've been using StackProf, which is great except that it doesn't work on
jruby.

On Wed, Feb 25, 2015 at 9:04 AM, Jonathan Rochkind <notifications@github.com

billdueber:
OK, so I [benchmarked](https://gist.github.com/billdueber/654cbbb30f40b1c4a50b) the unrolling of the ToField lambda/block and found a consistent speedup of 1-3% under MRI and maybe 7% under JRuby (but who knows what JRuby is optimizing away...). Even at 3%, that's a decent boost for a pretty minor duplication of code.

billdueber:
@barmintor Totally let this fall off my radar. I'll try to understand it all this weekend. Most of it (array-unrolling, frozen constants, pre-generating the list of interesting tags via `interesting_tags_hash.keys`) are no-brainers and I'll put them in. The loop unrolling doesn't seem to do as much for modern JRuby implementations, but a few percent here is a nice effect.

barmintor:
no problem @billdueber. I'm not sure what the current state of the difference would be, but 1) blacklight_marc merged my traject branch into master, and 2) we'll be looking at it for our catalog, so every % will help!

billdueber:
@barmintor Cool! I'm about to start some refactoring -- probably about time to start marking methods as `private` to make sure (i.e., enforce) that we have the same understanding of what the public interface is.

Anywhere in your code that you might be dipping into the internals more than I'd expect? I don't want to break you.

In any case, maybe I'll go crazy with the private methods, and then you can test to make sure I didn't break anything, and then I'll document them a little better in the YARD docs. Sound good?

barmintor:
Sure- there's a CI suite on blacklight_marc, just let me know when there's a branch you'd like to see exercised.

billdueber:
@barmintor Incorporated much of this in #123,  I didn't find any actual improvements in performance under real load (over a 30-second test); will continue to investigate.

barmintor:
Great, thanks Bill!

PR #104, opened 2015-05-27
Allow the user to set the reader to any enumerable

billdueber:
On master, the reader is created only via the combination of the
(required) reader_class and the (required) input io. This allows the
user to set the reader either via `indexer.reader=` or `settings['reader']`
to any enumerable they'd like.

A side effect of this is that the call to process need not take an
IO object, so specifying one (in code or on the command line) is
optional.

Added code to indexer.process, updated the warning about a missing IO
in command line, and added a couple tests using a simple Range as a
reader.


billdueber:
Not sure I'm catching all the command-line cases. Closing until I am.

PR #122, opened 2016-01-12
Be more forgiving about the record ID field

billdueber:
attn: @jcoyne. This seem like a reasonable way to handle it?

DebugWriter errors out when encountering a record without
an 'id' field (or, more generally, whatever is set in
`debug_writer.idfield`).

New behavior is to emit a warning on first encounter, and use as the record id 
`record_no_N` (where N is the record position).

The warning reads:

```ruby
"At least one record (##{context.position}) doesn't define field '#{@idfield}'.
All records are assumed to have a unique id. You can set which field to look in via the setting 'debug_writer.idfield'"

```

Tests added.

### DIFF
--- a/lib/traject/debug_writer.rb
+++ b/lib/traject/debug_writer.rb
@@ -32,14 +32,40 @@ require 'traject/line_writer'
 #       provide "output_file", "out.txt"
 #     end
 class Traject::DebugWriter < Traject::LineWriter
-  DEFAULT_FORMAT = '%-12s %-25s %s'
   DEFAULT_IDFIELD = 'id'
+  DEFAULT_FORMAT  = '%-12s %-25s %s'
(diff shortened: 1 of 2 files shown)


jcoyne:
Yes, this is great. Good job @billdueber !

billdueber:
Great. Once I get the problems with travis ironed out, I'll pull in these
changes and do a minor upgrade.

On Tue, Jan 12, 2016 at 3:25 PM, Justin Coyne <notifications@github.com>
wrote:

> Yes, this is great. Good job @billdueber <https://github.com/billdueber> !
>
> —
> <https://github.com/traject/traject/pull/122#issuecomment-171043649>.
>



-- 
Bill Dueber
Library Systems Programmer

PR #126, opened 2016-02-14
Allow nil values, empty fields, and deduplication

billdueber:
This adds three new settings (all of whose defaults reflect current behavior)

* `allow_nil_values` (default: false). Allow nil values to be sent on to the writer
* `allow_duplicate_values` (default: true). Allow duplicate values. Set to false to
  force only unique values.
* `allow_empty_fields` (default: false). Default behavior is that the output hash
  doesn't even contain keys for a `to_field` which doesn't produce any values.
  Set to `true` to pass empty fields on (with the value being an empty array)

Tests have been added for all the new behavior.

Updated the README and settings.md, and reorganized the latter as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,29 @@ after_processing do
 end
 ~~~
 
+## Allowing/disallowing repeating duplicate values
+
+By default, `traject` allows the same value to be added to the same field without
+restriction. You can use the setting `allow_duplicate_values = false` to
+disallow duplicate values (i.e., call `uniq!` on the set of values associated with
+a given field).
+
+## Keeping nil values
+
+`traject` defaults to throwing away any `nil`s that make it into your accumulator
+of values. The setting `allow_nil_values = true` will let `nil` values
+pass through.
+
+# Dealing with  empty fields
+
+Similary, by default `traject` completely ignores empty fields. You can
+change this with the setting `allow_empty_fields = true`, which will result
+in the output hash having a key for every field mentioned in a `to_field`
+statement, whether or not it has any values in it.
+
+Fields that are empty will have a value sent to the writer of an empty
+array (`[]`). Writers that need to special-case empty fields should do so in the
+writer class in question.
 
 ## Readers and Writers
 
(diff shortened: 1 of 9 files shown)


billdueber:
@jcoyne Can you take a look at this and then take it for a spin?

billdueber:
@jcoyne pinging you one more time to see if you have any time to run this instead of your patched version and see if it'll do what you need it to do. If you don't have time, I'll just make sure my tests are all in line and push out a new version.

jcoyne:
Sure, let me take a look.

jcoyne:
@billdueber I don't think the readme does a good job of saying where to set those values.  I tried 
```ruby
extract_marc('650', trim_punctuation: true, allow_empty_fields: true)
```
but that obviously wasn't it.

billdueber:
Ah, yeah. They’re settings-level things, so you set them like the other
settings.

settings do
  provide "allow_empty_fields", true
end

I can’t be turned on/off by field; is that a use case I’m missing? I was
assuming that if you have a store that needs empty fields or nil values,
you’ll have to deal with them by hand regardless of the field.
​

On Tue, Feb 23, 2016 at 9:06 PM, Justin Coyne <notifications@github.com>
wrote:

> @billdueber <https://github.com/billdueber> I don't think the readme does
> a good job of saying where to set those values. I tried
>
> extract_marc('650', trim_punctuation: true, allow_empty_fields: true)
>
> but that obviously wasn't it.
>
> —
> <https://github.com/traject/traject/pull/126#issuecomment-188012507>.
>



-- 
Bill Dueber
Library Systems Programmer

jcoyne:
This is fine and it seems to work, I'm just wondering if the README should be improved.

billdueber:
Yup. Will do. Thanks so much for the feedback.

On Tue, Feb 23, 2016 at 9:13 PM, Justin Coyne <notifications@github.com>
wrote:

> This is fine and it seems to work, I'm just wondering if the README should
> be improved.
>
> —
> <https://github.com/traject/traject/pull/126#issuecomment-188013842>.
>



-- 
Bill Dueber
Library Systems Programmer

jcoyne:
Thanks for your work on this @billdueber!!! It's great to be back on master.

PR #128, opened 2016-04-01
bugfix to regexp in specification parsing

marquis:
Besides fixing the regexp (s to \s), I added an "s" subfield to the existing spec parsing test.


marquis:
Intermingled my two changes, canceling to redo

PR #274, opened 2021-07-28
Provide a fallback for a missing second date

fjorba:
If the record 008 date states a range of dates (q) but the second date is
missing, date2_str is undefined, and it causes an exception:

 Exception: NoMethodError: undefined method `sub' for nil:NilClass
 [...]/vendor/bundle/ruby/2.7.0/gems/traject-3.6.0/lib/traject/macros/marc21_semantics.rb:336:in
 `publication_date'

Assign date2_str the same value than date1_str as a fallback.

### DIFF
--- a/lib/traject/macros/marc21_semantics.rb
+++ b/lib/traject/macros/marc21_semantics.rb
@@ -327,7 +327,11 @@ module Traject::Macros
       if field008 && field008.length >= 11
         date_type = field008.slice(6)
         date1_str = field008.slice(7,4)
-        date2_str = field008.slice(11, 4) if field008.length > 15
+        if field008.length > 15
+          date2_str = field008.slice(11, 4)
+        else
+          date2_str = date1_str
+        end
 
         # for date_type q=questionable, we have a range.
         if (date_type == 'q')


lib/traject/macros/marc21_semantics.rb
```diff
@@ -327,7 +327,11 @@ def self.publication_date(record, estimate_tolerance = 15, min_year = 500, max_y
...
         date1_str = field008.slice(7,4)
-        date2_str = field008.slice(11, 4) if field008.length > 15
+        if field008.length > 15
+          date2_str = field008.slice(11, 4)
+        else
+          date2_str = date1_str
+        end
 
         # for date_type q=questionable, we have a range.
         if (date_type == 'q')
```
billdueber, line 337: With your changes this should probably  check for `'q'` OR `date1_str == date2_str`.

billdueber (changes requested):
Thanks for catching this, and for the first-time PR!

A couple things I'd love to see if you can:
  * The "questionable" branch (`date_type == 'q'`) should probably not run if date1 == date2
  * Putting in a test would be wonderful

lib/traject/macros/marc21_semantics.rb
fjorba: Understood. Please allow me some time, as I don't have practice with tests, and I have to learn how to do them.

fjorba: I've done my best, please bear with me. This has been my (arduous) first test task.  I haven't been able to test the test itself, sorry; it may be something to do with an orphan `/usr/local/bin/rake` that cannot find its ruby interpreter (I'm also quite new dealing different ruby interpreters).  However, I expect the Marc record to be correct.

Anyway, to my understanding the condition it reads better this way:

```
 if date_type == 'q' and date1_str != date2_str
```
and that's why I've put it this way. However, I'm ready to listen to counter arguments, thanks

fjorba:
I think that my part is done. Are you expecting me to do anything else, @billdueber ?  Thank you

billdueber (approved):

billdueber:
Hey @fjorba this is looking great! One last problem -- `missing-second-date.marc` isn't legal in some way -- it's somehow indicating that there's a second record in the file, and there isn't.  It can be easily fixed with something like the following code, to force-write only one record. 

```ruby
writer = MARC::Writer.new('fixed-missing-second-date.marc')
writer.write(MARC::Reader.new('missing-second-date.marc').first)
writer.close
```

...and then `git mv fixed-missing-second-date.marc missing-second-date.marc`

fjorba:
I think that it was an extra newline at the end of the file. I'm sorry that I still cannot test the test myself; some day I'll fix my environment.

PR #294, opened 2024-02-06
Bug fix for the `#filing_version` logic

billdueber:
Derivation of the filing version assuming the first subfield in a field would hold content (e.g., `$a`) and thus failed when it held a pointer to a linking field (e.g., `$6 245-01`). Now do checks and modification of the first _alphabetic_ subfield. 

Fixed the code, added a test, changed CHANGES.md and upped the patch version.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,14 @@
 
 *
 
-*
+## 3.8.2
+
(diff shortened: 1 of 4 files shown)


billdueber:
The 9.1 failure is something about the CI environment.  See https://github.com/jruby/jruby/issues/4834 -- it was fixed in later versions.

PR #297, opened 2024-12-05
Stop testing on ancient versions of ruby

jcoyne:


jcoyne:
@billdueber that data was gathered 2 years ago.

PR #308, opened 2025-03-26
Update marc language and geographic maps from LOC

maxkadel:
Small updates to load_maps rake task to make it work with recent Ruby, and include date in language file


billdueber:
Thanks, @maxkadel! I've got a couple other minor changes I might want to fold in, and then I'll release a patch version.

PR #310, opened 2026-04-27
pass CI in ruby 3.4

jrochkind:
We have CI passing in ruby 3.4, basically just required adding some dependencies that were moved out of default ruby or otherwise required updating. 

I wanted to get passing in JRuby, but it's beyond me -- I do not use jruby. 

> Traject::ExperimentalNokogiriStreamingReader::with namespaces::floating path#test_0001_reads:
Java::OrgW3cDom::DOMException: NAMESPACE_ERR: An attempt is made to create or change an object in a way which is incorrect with regard to namespaces.

I did drop old jrubies from CI, and added only jruby 10.0, which is ruby 3.4 compatible. It isn't really feasible to support very old jrubies, jruby doesn't work that way.  Jruby 10.1 is also out supporting ruby 4.0, which we don't yet support, will have that in subseuqent ticket. 

@billdueber thoughts on what we should do with jruby?  Drop from CI or at least from mandatory passing of CI? No longer support Jruby?  You still use traject with jruby? Know of others who do?

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -19,8 +19,7 @@ jobs:
           - '3.2'
           - '3.3'
           - '3.4'
-          - 'jruby-9.3'
-          - 'jruby-9.4'
+          - 'jruby-10.0'
     name: Ruby ${{ matrix.ruby }}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


mksndz:
Thanks @jrochkind for this. I'm don't have the rights to do a formal review but this looks straightforward to me. I set up a development environment using this branch and Ruby `3.4` and things are all green, whereas I get the `base64` issue on `main`.

PR #311, opened 2026-04-27
avoid ruby 3.4 warning for "literal string will be frozen in the future"

jrochkind:
"literal string will be frozen in the future" warning for string literals that in future ruby version will be automatically frozen, but our tests modify. 

Fixed in our code. The way I fixed for `escape` method is the most unusual, but I think it's the only good way to do it?  Feedback welcome, @billdueber ?  Other places were fairly straightforward. 

Warnings remain for two dependencies: 1) fastxmlwriter (only used in a benchmark test), 2) another still remains from "yell" (will submit a PR to yell). 

- update context with frozen_string_literal default
- frozen_string_literal in experimental_nokogiri_streaming_reader
- frozen_string_literal in marc21_semantics.rb
- Fix frozen string warnings for 'escape' in Delimited writers. Not totally sure in what cases this is necessary in non-test code, but this seem safest based on it being necessary in test code

### DIFF
--- a/lib/traject/csv_writer.rb
+++ b/lib/traject/csv_writer.rb
@@ -5,7 +5,6 @@ require 'csv'
 # Use DelimitedWriter for non-CSV lines (e.g., tab-delimited)
 #
 #
-
 class Traject::CSVWriter < Traject::DelimitedWriter
 
   def initialize(*args)
@@ -26,6 +25,11 @@ class Traject::CSVWriter < Traject::DelimitedWriter
   # Let CSV take care of the comma escaping
   def escape(x)
     x = x.to_s
+
+    if x.frozen?
+      x = x.dup
+    end
+
     x.gsub! internal_delimiter, @eidelim
     x
   end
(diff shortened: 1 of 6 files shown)


jrochkind:
Pulled to yell at https://github.com/rsmdt/yell/pull/66

lib/traject/csv_writer.rb
```diff
@@ -26,6 +25,11 @@ def open_output_file
...
   def escape(x)
     x = x.to_s
+
+    if x.frozen?
+      x = x.dup
+    end
+
     x.gsub! internal_delimiter, @eidelim
     x
   end
```
mksndz, line 35: Why not just avoid any mutation in these escape methods?

```ruby
  def escape(x)
    x&.to_s&.gsub(internal_delimiter, @eidelim)
  end
```

jrochkind: Doing so would create a new string with every `gsub`, and creating new objects is often a performance issue in ruby. 

This is in a "hot" spot, called many times -- once per field (or even subfield?) if you are using these output methods. 

@billdueber has done some performance optimization in the past, so I figured best to keep it behaving similarly to before, not creating more new objects. (Except for the one it has to create if the input is frozen!).  

I did consider it, so we are following the same paths!

mksndz: OK, that makes sense - thanks for the explanation

mksndz:
All this seems fine! And TIL about https://docs.ruby-lang.org/en/3.1/String.html#method-i-2B-40

mksndz (approved):
👍

PR #313, opened 2026-04-27
Support and test on Ruby 4.0

jrochkind:
- remove bundler from gemspec
- Replace CGI.* methods removed from ruby 4.0 with URI.* methods
- test on ruby 4.0 and corresponding jruby 10.1


mksndz (approved):
👍

